PR #57, opened 2022-01-13
Alex input updates

lindsayad:
I uncommented the block deletion generation in order to be sure that my
new changes I just pushed to the interpolator PR work. Before my newest
commit on the interpolator PR this input would finish residual
evaluation when the blocks were deleted but would fail if the blocks
were not deleted. I also added block restriction to all the FV kernels;
we should probably add an integrity check in MOOSE to make sure that
kernel block restriction is always a subset of variable block
restriction.

Finally, the change that is important to get the nonlinear solve to
converge is to add the `-pc_factor_shift_type NONZERO` flag

### DIFF
--- a/msr/msfr/steady/compare_ns.i
+++ b/msr/msfr/steady/compare_ns.i
@@ -13,19 +13,23 @@
 # A coupled neutronics-coarse TH solution can be generated with
 # 'run_neutronics.i', saved in 'run_neutronics_ns_restart.e'
 
+# Numerical scheme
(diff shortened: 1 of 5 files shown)


lindsayad:
@GiudGiud these small changes with my current branch tip of `insfv-interpolator` and our current branch tip of `update-functors-2` make that input run quite well, and IMO the results look pretty nice for the mesh. As I said on slack the bypass fraction is roughly 5%. Signing off for the weekend!

PR #102, opened 2022-06-22
Add MultiApps Documentation to VTB

### LINKED ISSUE #101: Add MultiApps Tutorial Documentation
Add MultiApps tutorial into VTB documentation.

miaoyinb:
closes #101


miaoyinb:
@aabdelhameedd

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Type of MultiApp
+
+The `type` parameter describes the time/space-based execution of the child application. Various MultiApp `type`s are available to facilitate different possible solution schemes of multi-physics problems. Available `type`s include: [`FullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/FullSolveMultiApp.html), [`TransientMultiApp`](https://mooseframework.inl.gov/source/multiapps/TransientMultiApp.html), [`CentroidMultiApp`](https://mooseframework.inl.gov/source/multiapps/CentroidMultiApp.html), [`Stochastic Tools`](https://mooseframework.inl.gov/modules/stochastic_tools/index.html) , and [`LevelSetReinitializationMultiApp`](https://mooseframework.inl.gov/source/multiapps/LevelSetReinitializationMultiApp.html).  
+
+#### FullSolveMultiApp
+
+The [`FullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/FullSolveMultiApp.html) object performs a complete child application simulation during each execution which means that the child applications will be solved from the first-time step to the last time step every time it is called.
```
GiudGiud, line 17:
```suggestion
The [`FullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/FullSolveMultiApp.html) object performs a complete child application simulation during each execution which means that the child applications will be solved from the first-time step to the last time step every time it is called. This is typically done for steady state or eigenvalue parent-app simulations or for initialization purposes.
```

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
+
+ The [`Stochastic Tools`](https://mooseframework.inl.gov/modules/stochastic_tools/index.html) apps include [`SamplerFullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/SamplerFullSolveMultiApp.html) , [`SamplerTransientMultiApp `](https://mooseframework.inl.gov/source/multiapps/SamplerTransientMultiApp.html) , and [`PODFullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/PODFullSolveMultiApp.html) . [`SamplerFullSolveMultiApp`](https://mooseframework.inl.gov/source/multiapps/SamplerFullSolveMultiApp.html)  and [`SamplerTransientMultiApp `](https://mooseframework.inl.gov/source/multiapps/SamplerTransientMultiApp.html) create full solve type, and transient type child application, respectively.
+
+#### MultiApp Classes in `Level Set App`
+
+The [`LevelSetReinitializationMultiApp`](https://mooseframework.inl.gov/source/multiapps/LevelSetReinitializationMultiApp.html) method is used for front tracking problems, and its mathematical formulations represents the structural solid-void boundary.
```
GiudGiud, line 86:
```suggestion
```
too specific imo

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
+
+The `execute_on` parameter specifies when a child application is to be executed, in relation to the parent application timesteps. The most common options are to execute the child application at the beginning of the time step, at the end of the time step, or at the initial condition of the entire simulation. By default, child application execution will occur at the beginning of time-step.  In the previous example, at every time step, the main application will be executed first, then the child applications will be executed at the end of the time step. Some common options of `execute_on` are summarized below:
+
+  1. `TIMESTEP_BEGIN`: Launch child application at the beginning of the time step
+
+  2. `TIMESTEP_END`: Launch child application at the end of the time step
+
+  3. `INITIAL`: Launch child application only at the initial condition to set up the initial conditions for the solve
+
+  4. `NONLINEAR`: This option has limited applications and is used to inject the coupling information into the middle of the Newton solve which will slow down the newton convergence but could speed up the conversion of the entire system.
```
GiudGiud, line 120:
```suggestion
  4. `NONLINEAR`: This option has limited applications and is used to inject the coupling information into the middle of the Newton solve. It will likely slow down the newton convergence, as the contribution to the Jacobian are not considered, but could speed up the convergence of the entire system.  
```

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
+#### Finer Timesteps in Child Applications
+
+Applications may use different time step sizes by allowing `subcycling`. The subcycling option is useful for allowing different timescales. For instance, a relatively slow physics application may use bigger time steps (e.g. 10 sec) whereas a rapidly changing physics may use relatively small-time step (e.g. 0.1 sec). Thus, the first app takes one time-step of 10 sec while the other app takes 100 time-steps of 0.1 sec to reach the same end time as the first application.
+
+!media media/resources/advsubcycling.png
+      style=display: block;margin-left:auto;margin-right:auto;width:80%;
+      id= subcycling
+      caption=Schematic drawing of the sub-cycling concept.
+
+ [subcycling] illustrates a transient Multiphysics simulation in which Griffin, Bison, and SAM are coupled at different timescales using the subcycling option. When performing sub-cycling, transferred auxiliary variables on child applications may be interpolated between the start time and the end time of the main app with the `interpolate_transfers` parameter. When using the TransientMultiApp type, the `allow_sub_cycling` parameter can be used to invoke subcycling.
```
GiudGiud, line 131:
```suggestion
 [subcycling] illustrates a transient Multiphysics simulation in which Griffin, Bison, and SAM are coupled at different timescales using the subcycling option. When performing sub-cycling, transferred auxiliary variables on child applications may be interpolated between the start time and the end time of the main app with the `interpolate_transfers` parameter. When using the TransientMultiApp type, the `sub_cycling` parameter can be used to invoke subcycling.
```

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
...f the parent application geometry by using the 'positions' input parameter. Positions are simply a list of 3D coordinates that describe the offset of the child application's origin from the parent application's origin. If no positions are provided, the child application will be launched at position is `0 0 0` within the parent application. Notice that in the example below, the number of child applications to be invoked is not explicitly defined but rather is inferred from the number of positives given. The example shows five different positions, thus, five independent versions of the child application `BisonApp` will be created and each of them will be associated with a different position within the parent application domain. The user may also provide a position file instead of explicitly providing the positions in the MultiApp block; this feature is useful when the user wants to invoke thousands of child application as the separate input file can contain all this information without cluttering the main input file. A common use of positions in nuclear reactor analysis is the launching of individual fuel rod (BISON) simulations at various fuel rods within a reactor assembly or core.
```
GiudGiud, line 135:
```suggestion
Child applications can be launched at various spatial positions of the parent application geometry by using the 'positions' input parameter. Positions are simply a list of 3D coordinates that describe the offset of the child application's origin from the parent application's origin. If no positions are provided, the child application will be launched at the position `0 0 0` within the parent application. Notice that in the example below, the number of child applications to be invoked is not explicitly defined but rather is inferred from the number of positions given. The example shows five different positions, thus, five independent versions of the child application `BisonApp` will be created and each of them will be associated with a different position within the parent application domain. The user may also provide a position file instead of explicitly providing the positions in the MultiApp block; this feature is useful when the user wants to invoke thousands of child application as the separate input file can contain all this information without cluttering the main input file. A common use of positions in nuclear reactor analysis is the launching of individual fuel rod (BISON) simulations at various fuel rods within a reactor assembly or core.  
```

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Other Options
+
+ For the MOOSE Apps: `FullSolveMultiApp`, `TransientMultiApp`, and `CentroidMultiApp`. There are various common optional parameters such as `app_type`, `bounding_box_inflation`, `bounding_box_padding`,  `catch_up`, `cli_args`, `cli_args_files` , `clone_master_mesh` , `detect_steady_state`,  `global_time_offset` , `keep_solution_during_restore`, `library_name`, `library_path`,`output_in_position`, and `execute_on`.
+
+ Other parameters are allowed only for specific MultiApps type. For example, `positions` is an optional parameter for FullSolveMultiApp and TransientMultiApp, but it is not a paramter for CentroidMultiApp. Generally, these optional parameters were developped to facilitate flexiabile multi-physics simulations.  The `no_backup_and_restore` optional parameter is useful when doing steady-state Picard iterations where we want to use the solution of previous Picard iteration as the initial guess of the current Picard iteration.
```
GiudGiud, line 171:
```suggestion
 Other parameters are allowed only for specific MultiApps type. For example, `positions` is an optional parameter for FullSolveMultiApp and TransientMultiApp, but it is not a parameter for CentroidMultiApp. Generally, these optional parameters were developed to facilitate flexible multi-physics simulations.  The `no_backup_and_restore` optional parameter is useful when doing steady-state Picard iterations where we want to use the solution of previous Picard iteration as the initial guess of the current Picard iteration.
```

GiudGiud (commented):
first pass. This is incredibly detailed, we ll definitely want to link to it from the framework website

in Fig 2-2 we would want the left bottom to be split as well

in the diagram P is power, can be confused for pressure
h is heat transfer coefficient, doesnt jump to mind either
delayed neutrons -> delayed precursors

the vtb_examples diagrams are incomplete:
MSR: fission source is sent as well, not just power
PBFHR: pebble model is roughly the same as PBMR, Tbc or T_matrix is passed down
SFR: the inlet temperature is not passed currently, it's set

GiudGiud:
@miaoyinb I m going to use these figures for the ANS summary we are both on if that's ok

doc/content/resources/multiapps_chps/chp_3_applications.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### VTB-HTGR
+
+In the High-Temperature Gas-cooled Reactor ([HTGR](htgr/index.md)) category, a Pebble-Bed Microreactor ([PB-MR](pbmr/index.md))) model is available. Griffin is the parent application governing neutronics simulations, and MOOSE Heat Conduction and Thermal Hydraulics modules child applications calculate the heat transfer in solid (matrix + fuel) and fluid (coolant) regions, respectively. A special approach used in this example is the use of a CentroidMultiApp as the grandchild app to calculate the representative temperature profile within single TRISO particles using Heat Conduction Modules. This is an example of using MultiApps to handle similar physical phenomena (thermal here) at different space scales.
```
GiudGiud, line 88:
```suggestion
A Pebble-Bed Microreactor model is available on the VTB using a three-level MultiApps structure. Griffin is the parent application governing neutronics simulations, while Pronghorn solves a homogenized porous media flow problem. A CentroidMultiApp is then used to calculate the representative temperature profile within pebbles and TRISOs. This is an example of using MultiApps to handle similar physical phenomena (thermal heat transfer here) at different space scales. 
```

both child input files use Pronghorn not the modules. There's a lot of modules code in them but you cant run them without prognhorn because of the matprops

miaoyinb:
> @miaoyinb I m going to use these figures for the ANS summary we are both on if that's ok

Should be fine. Just need to confirm with @eshemon.

miaoyinb:
> 
> in Fig 2-2 we would want the left bottom to be split as well
> 
@aabdelhameedd Can you make a new Fig. 2-2 per @GiudGiud's comment. Thanks.

miaoyinb:
> 
> in the diagram P is power, can be confused for pressure h is heat transfer coefficient, doesnt jump to mind either delayed neutrons -> delayed precursors
> 
I will replace $h$ with $htc$. For P, I guess I have to explain it in the caption.

miaoyinb:
> the vtb_examples diagrams are incomplete:
> MSR: fission source is sent as well, not just power
> PBFHR: pebble model is roughly the same as PBMR, Tbc or T_matrix is passed down
> SFR: the inlet temperature is not passed currently, it's set

Will fix these.

aabdelhameedd:
> > in Fig 2-2 we would want the left bottom to be split as well
> 
> @aabdelhameedd Can you make a new Fig. 2-2 per @GiudGiud's comment. Thanks.

@GiudGiud I am not sure that I clearly understanding  the required modification. Do you mean remove the left bottom child applications under MultiApp(3)? Or do you mean I should have sequence like first, second, third, ..etc that represents the sequence of execution?

GiudGiud:
In figure 2.2 you have two grey blocks on the right and left labelled first and second. Then on the left in the first block, you have another set of MultiApps (grandchild) inside the same child app. There are two issues:
- the number of ranks jumps from 6 (or 4 on the right side) back to the initial one (12). I have not checked but I highly doubt it would work this way
- MultiApp (3) and (4) are confusing. It is technically possible for the two child apps to have a different MultiApp, but I dont think this is the right example.
I think we should have (3) (4) on both subapps instead.

aabdelhameedd:
> In figure 2.2 you have two grey blocks on the right and left labelled first and second. Then on the left in the first block, you have another set of MultiApps (grandchild) inside the same child app. There are two issues:
> 
> * the number of ranks jumps from 6 (or 4 on the right side) back to the initial one (12). I have not checked but I highly doubt it would work this way
> * MultiApp (3) and (4) are confusing. It is technically possible for the two child apps to have a different MultiApp, but I dont think this is the right example.
>   I think we should have (3) (4) on both subapps instead.

Thank you for your comments. For the first comment, we had a discussion on Slack-VTB about this figure few months ago, what I understood from Cody's reply is that the “Multiapp” does indeed have access to the full set of MPI ranks because we only execute one Multiapp at a time ". Please let me know if I misunderstood Cody's reply on Slack?

GiudGiud:
Nested MultiApp dont get the full set of MPI rank from the grandparent. 
MultiApps get access to the full set of MPI rank from their parent.
@permcody correct me if I m wrong please

aabdelhameedd:
> @GiudGiud Is it possible that more than one "nested" MuliApp run at any time on the system?

permcody:
No, each multiple runs separately, but a single multiapp may contain
several subapps, which are scheduled simultaneously.

GiudGiud:
Nested MultiApps may run at the same time.
The processes are spread at each level. But each MultiApp (child level) being ignorant of the fact that it is a MultiApp it cant synchronize with its siblings for running its own MultiApps (grandchild level).

But at this point it's getting complicated and only confuses the reader about the two main points, which are made about the CHILD multiapps. Thoughts on not mentioning this?

miaoyinb:
> Nested MultiApps may run at the same time. The processes are spread at each level. But each MultiApp (child level) being ignorant of the fact that it is a MultiApp it cant synchronize with its siblings for running its own MultiApps (grandchild level).
> 
> But at this point it's getting complicated and only confuses the reader about the two main points, which are made about the CHILD multiapps. Thoughts on not mentioning this?

Thanks for the clarification. I agree it's getting complicated. 

Maybe we should try to make the `Parallel Execution` part briefer? For example, just to state the fact @GiudGiud mentioned above? Let me know if you two have any suggestions, @GiudGiud @aabdelhameedd.

miaoyinb:
@GiudGiud @permcody 
We updated the parallel related figure and related description. Let us know what you think. Thanks.

doc/content/resources/multiapps_chps/chp_3_applications.md
```diff
@@ -0,0 +1,122 @@
...
+### MultiApp Reactor Examples on VTB
+
+The examples available on the Virtual Test Bed site adopt a number of different coupling approaches based on the specific reactor type and phenomena of focused interest. We briefly summarize some examples to show readers the flexibility in using MultiApps (see [vtb_examples]).
+
+!media media/resources/vtb_examples.png
+      style=display: block;margin-left:auto;margin-right:auto;width:100%;
+      id=vtb_examples
+      caption=Hierarchies of MultiApps for VTB examples.
+
+#### VTB-MSFR
```
GiudGiud, line 81: for msfr, htgr, sfr and pb-fhr we have more than one model.
- let's link to the model in particular rather than the index page
- let's be more precise when referring to the model. For example instead of MSFR example, say MSFR core multiphysics simulation

doc/content/resources/multiapps_chps/chp_3_applications.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### VTB-HP-MR
+
+The Heat-Pipe Microreactor ([HP-MR](mrad/index.md)) uses a two-level MultiApps approach with a BISON parent application to govern the thermal physics within all the solid reactor components except for heat pipes, and a Sockeye child app to deal with heat pipe performance. A placeholder is added for future adoption of Griffin as parent app for neutronics simulation. When Griffin is integrated as parent, BISON and Sockeye will become child and grandchild applications, respectively.
+
+#### VTB-PBFHR
+
+The Pebble Bed Fluoride-Salt-Cooled High-Temperature Reactor ([PB-FHR](pbfhr/index.md)) uses an approach similar to the HTGR example. Unlike the MSFR model in VTB where fuel is dissolved in the salt, the fuel of PBFHR has solid form so that delayed neutron information no longer needs to be handled by fluid dynamics application. However, the thermal physics inside the fuel pebbles need to be simulated by an additional grandchild Heat Conduction application.
```
GiudGiud, line 95:
```suggestion
The Pebble Bed Fluoride-Salt-Cooled High-Temperature Reactor ([PB-FHR](pbfhr/index.md)) uses an approach similar to the HTGR example. Unlike the MSFR model in VTB where fuel is dissolved in the salt, the fuel of PBFHR has a solid form. Therefore delayed neutron information no longer needs to be handled by fluid dynamics application. However, the thermal physics inside the fuel pebbles need to be simulated by an additional grandchild Heat Conduction application.
```

doc/content/resources/multiapps_chps/chp_3_applications.md
```diff
@@ -0,0 +1,122 @@
...
...he HTGR example uses MultiApps for non-multiphysics coupling. As mentioned early in this chapter, MultiApps are used in this example to connect a similar physical phenomenon (i.e., heat conduction) at two different space scales. In the HTGR example, the macroscopic reactor components involved in neutronics and thermal hydraulics simulations such as fuel and coolant are meter level, while single TRISO particles with fuel pebble are only millimeter level. Due to the complex multilayered structure of a TRISO particle, the particle needs to be separately simulated to capture its internal maximum temperatures. It is impractical to fully couple them as that would require an ultra fine mesh to capture thousands of TRISO particles. Instead, using a CentroidMultiApp, a virtual representative TRISO particle is simulated for each macroscopic fuel mesh element, providing representative local maximum fuel temperature information. This example shows the advantage of using MultiApps to study similar (or dissimilar) physical phenomena happening at very different space scale. Using similar approaches, it is possible to connect engineering scale simulations with mesoscale or even microscale physics.
```
GiudGiud, line 107: link to the example. 
there are 3 inputs now for HTGR, 2 have multiapp setups

doc/content/resources/multiapps_chps/chp_4_workflow.md
```diff
@@ -0,0 +1,118 @@
+# 4. Workflow to Establish a MultiApps Simulation
+
+## From Physics to Models
+
+### Identification of Physical Phenomena and Their Domains
+
+The first step when designing a MultiApps based simulation is to identify the physical phenomena that needs to be modeled, and their spatial and time domains.
```
GiudGiud, line 7:
```suggestion
The first step when designing a MultiApps based simulation is to identify the physical phenomena that need to be modeled, and their spatial and time domains.
```

doc/content/resources/multiapps_chps/chp_4_workflow.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Determine MultiApps Hierarchy
+
+The required MOOSE apps (solving for physical phenomena) are interconnected by the data that needs to be exchanged. The MultiApps hierarchy needs to be established to finalize the design of the MultiApps system. There are generally many options for working hierarchies (i.e. not unique), but some hierarchies may provide better computational efficiency. One major limitation that restricts the design of the MultiApps hierarchy is that only parent-child data exchange is allowed in the current MOOSE [`Transfers`](https://mooseframework.inl.gov/syntax/Transfers/index.html) system. Namely, sibling-sibling or grandparent-grandchild data exchange is not yet supported.
```
GiudGiud, line 46: rework this to say sibling-sibling is su[ported

doc/content/resources/multiapps_chps/chp_4_workflow.md
```diff
@@ -0,0 +1,118 @@
...
...r conserves the integrated power, leveraging $L^2$ projection to minimize difference between the integrated power. For the temperature variable from heat conduction and thermal hydraulics applications, as well as the displacement variables from tensor mechanics application, `MultiAppInterpolationTransfer` is used to ensure proper values can be obtained when meshes are different. Because the thermal physics application and the tensor mechanics applications use an identical mesh, field variables such as temperature and displacements are transferred by `MultiAppCopyTransfer`. The data transfer between the thermal physics application (BISON) and the thermal hydraulics (SAM) application is special. Although both applications use 2D-RZ axisymmetric meshes, the applications are oriented in different coordinate systems. In BISON, the convention is that the R mesh exists in the XZ plane, while in SAM, the convention is that R mesh exists in the XY plane. `MultiAppCoordSwitchNearestNodeTransfer` (a SAM object) performs transfers with coordinate system conversion along with the nearest node transfer. Native support is also present in MOOSE for most coordinate-transformations during transfers.
```
GiudGiud, line 98: to conserve power, one has to use the two postprocessors. 

a L2 projection in general will not conserve power (L1 norm), unless strict conditions are met

I would cut this paragraph in sections for each transfer for readability
also add a code listing in between  each section

doc/content/resources/multiapps_chps/chp_5_multiapps.md
```diff
@@ -0,0 +1,212 @@
...
+    type = TransientMultiApp
+    app_type = BisonApp
+    input_files = 'Bison_input.i'
+    positions = '0   0   0
+                 0.5 0.5 0
+                 0.9 0.9 0'
+  []
+[]
+
+The minimum time step over the main and all child applications will be used, by default, unless `sub_cycling` is enabled. If sub_cycling is enabled this will allow faster calculations, as the child applications will be able to perform multiple time steps per execution.
```
GiudGiud, line 63:
```suggestion
The minimum time step over the main and all child applications will be used, by default, unless `sub_cycling` is enabled. If sub_cycling is enabled this will allow faster calculations, as the parent app will not be forced to use the smaller time step from the child app.
```

GiudGiud (changes requested):
I like the new figure. Should be good after these comments are addressed

eshemon:
Thanks again @GiudGiud for your very helpful review and suggestions!

permcody:
This is correct. However since we usually use the full set of ranks, every
level will has access to the full set. I’m not sure what happens when an
intermediate level has a down-selected set. I assume that this sets the max
for the subsequent level(s), which could make some tree configurations
suboptimal.

PR #109, opened 2022-07-14
Fixed gold file for Sockeye test

joshuahansel:
A Sockeye bug was discovered where the capillary limit was
computed incorrectly for HeatPipeConduction for vertical
heat pipes.

Refs #0


GiudGiud (approved):
thanks for the patch!

PR #185, opened 2023-01-07
Update VPP for pin power tally in LFR assembly problem

### LINKED ISSUE #184: Update the pin power tally vectorpostprocessor in the Griffin standalone model for the LFR problem
In the recent PR #169  for Griffin LFR model, the pin powers were tallied using a somewhat outdated vectorpostprocessor `HexagonalGridVariableIntegral` which requires detailed geometry in its input parameters. It is recommended to use the newly developed and recommended workflow directly compatible with the Reactor Module for best practices by using `ExtraIDIntegralVectorPostprocessor` which uses just extra element integer IDs to identify different pins so that geometry information is not needed.

pak1sol:
This closes #184.


pak1sol:
Hi @GiudGiud. Could you please help review and approve this? This PR is for updating the VPP type for the pin power tally.

GiudGiud (approved):
looks great. Thanks for the improvement!

PR #186, opened 2023-01-19
Move apps to github.inl.gov

loganharbour:
May need recipe changes


GiudGiud (dismissed):

loganharbour:
@GiudGiud this is good to go. I setup redirects on github.inl.gov for idaholab/app -> ncrc/app so that the old versions of apps will work

GiudGiud (approved):

PR #208, opened 2023-03-20
Remove use of ad_indexing_type in tests spec

lindsayad:
Refs idaholab/moose#23665


GiudGiud (approved):
removed upstream.
No failures on this because all codes were built with global indexing type

PR #246, opened 2023-06-19
Add `equivalence_grid_variables` to VTR griffin input file

shikhar413:
Addresses #245


GiudGiud (approved):
thanks for the patch!

PR #277, opened 2023-09-06
Update SAM model deprecated params

### LINKED ISSUE #276: Update deprecated params in SAM input files
Deprecated params/objects in SAM VTB input models should be updated, in anticipation of some code cleanup that will occur upstream in SAM.

travismui:
Closes #276 

Refs [SAM/SAM!804](https://git-nse.egs.anl.gov/SAM/SAM/-/merge_requests/804), where the deprecated `FunctionFluidProperties` object is removed from SAM and updated in the MHTGR model.


travismui:
Sorry @GiudGiud I need to revisit the MHTGR model, looks like this update breaks the model although it passes input check. This is not an urgent PR, I'll ping you when it's ready again for review.

travismui:
> Sorry @GiudGiud I need to revisit the MHTGR model, looks like this update breaks the model although it passes input check. This is not an urgent PR, I'll ping you when it's ready again for review.

OK, we should be good to go now, the MHTGR model was updated and runs, but it's still heavy so I will leave it as a syntax check. An overhaul of the model may come in a future PR. Thanks @zhieejhia93 for help with fixing the model and removing unused params.

GiudGiud (approved):
looks good. thanks for the deprecation patch!

PR #344, opened 2024-01-13
Update SAM submodule & syntax updates, refs #0

travismui:
@GiudGiud Updated SAM submodule in this and also fixed some models (including HTTF ring model). Couldn't push to the existing branch, so opening a new PR here.

Supersedes #321


GiudGiud (approved):
looks good! Thanks for taking care of that model

PR #378, opened 2024-04-11
fix null transient issue for MRAD

### LINKED ISSUE #377: MRAD HPMR does not hold a null transient
## Bug Description
The null transient in `MRAD/transient_null/HPMR_dfem_griffin_trN.i` does not hold a null, it constantly changes with the timestep, despite restarting from a converged steady state.

## Steps to Reproduce
Run the steady state HPMR files, then run the null transient input.

Pretty sure this comes because of the change to the transfers in the various versions of the code. The transfers need to be updated.

## Impact
Less confidence in HPMR model. Not a roadblock for my particular work, but of course we want the best results possible.

Tag @miaoyinb @nstauff

miaoyinb:
closed #377


GiudGiud (approved):
looks good thanks for the fix

PR #389, opened 2024-05-14
Made some fixes for Sockeye changes

joshuahansel:
A Sockeye change was made that caused some solution differences. Tests using the Sockeye executable have been fixed. We'll see if others come up in the multi-app testing.

Also, addressed a few existing deprecated usages.


GiudGiud (approved):
will merge when undrafted

I ll update DW submodule later, we ll live with the failure for now

PR #420, opened 2024-08-21
Increase test max time

### LINKED ISSUE #419: Test case timing out preventing devel to main merges
## Bug Description

The following case: virtual_test_bed/msr/msfr/plant/steady.core_fluids_only has been consistently timing out on MOOSE modules testing on devel to main merges. This causes the changes from reaching main. Testing of other apps against VTB, test against main and changes that have already been merged in to devel are needed to get testing to pass on those apps.

## Steps to Reproduce

Look at testing, fails semi-consistently with a timeout at 500 seconds. Fix is to just increase max_time in the test spec file to prevent timeouts.

## Impact

Prevents other app testing to pass.

gambka:
Closes #419.


gambka:
@GiudGiud

GiudGiud (approved):

PR #498, opened 2024-12-19
Control LFS includes from a config

### LINKED ISSUE #499: Control LFS includes via a config file
## Feature Description
Adding/removing LFS files for models currently requires a CIVET recipe update. This can be painful and isn't easy for non-privileged users. Control it with a version controlled config instead.

## Design
Add `civet/<app>.yaml` for each application. Have a list with a key `lfs_includes` which controls which LFS files are pulled.

## Impact
Easier model development workflow, reduces recipe burden.

loganharbour:
Closes #499

### DIFF
--- a/.civet/README.md
+++ b/.civet/README.md
@@ -1,0 +1,7 @@
+# .civet
+
+Contains configuration files that CIVET uses in CI.
+
+## Syntax
+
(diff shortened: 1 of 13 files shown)


GiudGiud (approved):
I imagine we ll need a recipes/scripts change but this looks good

PR #510, opened 2025-01-13
General automatic submodule udpate

pbehne:
New submodules for all apps


GiudGiud (approved):
All good. Thanks for doing this!

Issue #545, opened 2025-03-21
Remove unused inputs from PatternedHexMeshGenerator


shikhar413:
## Bug Description

Reactor module was updated to fix a bug with unused inputs used in PatternedHexMeshGenerator. VTB tests need to be updated to remove these unused inputs now

## Steps to Reproduce

See failures in https://civet.inl.gov/job/2766245/

## Impact

VTB test failures in Griffin

PR #546, opened 2025-03-21
Remove unused inputs from PatternedHexMeshGenerator

shikhar413:
Refs #545


GiudGiud (approved):
Thanks for the patch!

PR #549, opened 2025-03-26
Update graphite grade and regold tests

gambka:
Refs #537.


gambka:
@elementx54 for your reference.

gambka:
These three regolds pass just fine locally. It must be architecture dependent. I'll try regolding on HPC/Linux as that is what the test boxes use.

GiudGiud (approved):
thanks for the patch!

PR #558, opened 2025-04-17
Disabled ifelse extension

### LINKED ISSUE #557: Disable ifelse extension in MooseDocs
The `ifelse` extension is currently casing the `VTB Documentation` recipe to fail because the use of capabilities within the VTB is not appropriate. This is because the VTB does not use just a single application. The `ifelse` extension in the doc system should be disabled.

pbehne:
Closes #557.


pbehne:
Pronghorn failures are unrelated.

PR #604, opened 2025-06-17
Fix use of scaling parameters in msre th input

lindsayad:
Refs failures on [ncrc/pronghorn#1299](https://github.inl.gov/ncrc/pronghorn/pull/1299)


GiudGiud (approved):
good fix!

I think the new "shouldCreateVariable" API should catch this mistake. Is that what happened with Pgh & VTB?

lindsayad:
It caught the VTB mistakes yea! I only brought up Pronghorn because VTB testing on Pronghorn is failing because of this

Issue #678, opened 2025-10-13
Modify/add tests with tracking Griffin run times in mind


jthano:
## Feature Description
The VTB models can be useful tests to help track Griffin run times over time. This can help notify developers if some change in Griffin causes significant slow down. It can also help catch input file parameter changes that might lead to performance degradation. 

## Design
The purpose of this issue is to cover minor additions or modification to inputs and test files so that existing models can help with run time tracking. For example, slight parameter changes to make a VTB model more computationally expensive on purpose.

PR #679, opened 2025-10-13
add hpc_tests to microreactor model to start tracking griffin run times

jthano:
refs #678

### DIFF
--- a/microreactors/drum_rotation/hpc_tests
+++ b/microreactors/drum_rotation/hpc_tests
@@ -19,6 +19,7 @@
     input = empire_2d_CD_coarse.i
     cli_args = '--mesh-only'
     executable_pattern = 'griffin*'
+    group = 'griffin_performance'
   []
   # this test is primarily meant to help keep track of Griffin run times
   [neutronics_eigenvalue]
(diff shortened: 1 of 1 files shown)


microreactors/drum_rotation/hpc_tests
```diff
@@ -0,0 +1,73 @@
...
+    cli_args = '--mesh-only'
+    executable_pattern = 'griffin*'
+    prereq = drum_mesh
+    group = griffin_performance
+  []
+  [coarse_mesh]
+    type = RunApp
+    input = empire_2d_CD_coarse.i
+    cli_args = '--mesh-only'
+    executable_pattern = 'griffin*'
```
loganharbour:
```suggestion
    cli_args = '--mesh-only'
    executable_pattern = 'griffin*'
    group = 'griffin_performance'
```

microreactors/drum_rotation/hpc_tests
```diff
@@ -0,0 +1,74 @@
...
+    group = 'griffin_performance'
+  []
+  # this test is primarily meant to help keep track of Griffin run times
+  [neutronics_eigenvalue]
+    type = CSVDiff
+    input = neutronics_eigenvalue.i
+    cli_args = "Executioner/richardson_rel_tol=1e-6 Outputs/csv=true
+                Outputs/exodus=false
+                Outputs/file_base=neutronics_eigenvalue_heavy
+                TransportSystems/sn/NPolar=3
```
YaqiWang: I changed these two params to 2 and 6 in my PR. I guess that should be enough for a performance test. If my PR goes in first, we will not need the new gold files. Or remove these two params here and the new gold file. I certainly can change this after it is merged, but thought that we may want to have a slightly cleaner history.

PR #723, opened 2025-12-11
Griffin update to its current devel

### LINKED ISSUE #722: Some trivial regoldings are needed due to a Griffin initial guess change
## Reason
A change in Griffin on the initial guess for transport update is made into Griffin devel branch: https://github.inl.gov/ncrc/griffin/pull/2855. It affects several tests in VTB using DFEM-SN scheme and not having sufficient convergence for faster test. These tests need a trivial regolding.

## Design
Will manually update Griffin submodule to its devel hash and add the new gold files. Those files are in CSV so we can see the value changes.

## Impact
This is only affecting tests. If the convergence is fully, the change in Griffin should not affect the model results.

YaqiWang:
With some trivial regoldings. Closes #722.

### DIFF
--- a/msr/cnrs/s03/cnrs_s03_ns_flow.i
+++ b/msr/cnrs/s03/cnrs_s03_ns_flow.i
@@ -18,15 +18,6 @@ geometric_tolerance = 1e-3
 cavity_l = 2.0
 lid_velocity = 0.5
 
-[UserObjects]
-  [rc]
-    type = INSFVRhieChowInterpolator
-    u = vel_x
-    v = vel_y
-    pressure = pressure
-  []
-[]
-
 [Mesh]
   type = MeshGeneratorMesh
   block_id = '1'


GiudGiud:
These changes are likely from the same cause as the changes in this PR (#723)
But they were not run on this PR, because they are are HPC tests

The logs and gold files can be found in the virtual test bed project on HPC. They are copied after a run. Do you have access?

GiudGiud:
> Is it under a shared folder? I am not aware of it, so possibly do not have access.

yes where the shared folders usually are. I think it might be projects/virtual_test_bed/
Can you please check? Else @pbehne should have access.

pbehne:
I have access (you can't see /projects/virtual_test_bed until you cd into it). Let me know if you want me to open a regold PR.

GiudGiud:
It'd be best to do it now, while we know what changed the results.
I cant access HPC outside the country though

PR #728, opened 2025-12-18
Regold griffin results.

pbehne:
Regolds some griffin results as discussed in #723.


YaqiWang (approved):
This looks reasonable to me. The drum_rotation runs might have slightly different CPU times that are included in the markdown file. But that could be in the range of stochastic fluctuations. Sorry about not capturing this in my PR.

PR #744, opened 2026-01-27
Improve submodule updater robustness, HPC logic, and PR metadata

pbehne:
This PR updates submodule_updater.py to make submodule updates safer, more transparent, and better aligned with HPC workflows.

Changes:

- Add HPC vs current commit comparison: warn and skip when the current submodule is ahead of the HPC version.
- Sync submodules to the superproject immediately after resetting to the base branch.
- Handle inaccessible submodules gracefully:
  - Skip submodules that fail fetch with a warning.
  - Skip submodules that fail git submodule update during the post‑reset sync.
- Prepend the PR description with the actual command used, masking the config path as config.yaml, and print the same header to the console.

Notes:

The command header reflects the real invocation, with --config-file redacted to avoid leaking local paths.

Refs #0

### DIFF
--- a/scripts/submodule_updater.py
+++ b/scripts/submodule_updater.py
@@ -235,20 +235,21 @@ if __name__ == "__main__":
             hpc_sha = submodule_sha1[submodule.name]
             if hpc_sha != current_sha:
                 try:
-                    # This next line will trigger an exception if hpc_sha is NOT an ancestor of current_sha
-                    sm_git_repo.git.merge_base('--is-ancestor', hpc_sha, current_sha)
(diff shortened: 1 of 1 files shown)


sfr/subchannel/multiple_SCM_assemblies/19assemblies/fuel_assembly_1.i
GiudGiud: rebasing on main should remove these changes

GiudGiud:
Thanks I will merge this after using it on the next submodule update!

PR #755, opened 2026-02-16
Fix KRUSTY Deprecated "temp" Parameter

### LINKED ISSUE #754: Fix Global temperature and Deprecated temp Parameters for BISON
## Feature Description

BISON's testing failed to pass input file check due to a change in a deprecated parameter from [MOOSE #32209](https://github.com/idaholab/moose/pull/32209). Both a global `temperature = ` parameter and local block `temp = ` parameter were set. The PR made `temp` a deprecated parameter. The global parameter was then used as the new parameter was now `temperature`. No error was given that both `temperature` and `temp` were set.

## Design

Change the `temp` parameter to `temperature` locally. This will both remove the deprecation and provide the correct local priority value.

## Impact

This highlights the behavior of global params in connection to deprecation. This will fix the BISON test of VTB.

elementx54:
Switched KRUSTY BISON input UMo Thermal model `temp` parameter that was deprecated to the new `temperature` parameter. This local parameter provides the correct variable to the `HeatConductionMaterial` object instead of using the global `temperature` variable which exists on a different block.

Closes #754


GiudGiud:
@elementx54 the test failure is because the BISON submodule needs to be updated too

elementx54:
The change was in MOOSE, so do I update the submodule for BISON and MOOSE?

PR #757, opened 2026-02-17
Update Griffin MSR Coupled Region Depletion Tests with New Syntax #756

### LINKED ISSUE #756: Update Griffin MSR Multi-Region Depletion Tests to Use Current Capabilities
## Reason

The current Griffin multi-region depletion tests found in `virtual_test_bed/msr/generic_msr/depletion/` rely on capability from the Griffin `experimental` branch which is not maintained by the Griffin developers and will be deleted. These tests are also not tested by the VTB since they require the `experimental` branch. These tests should be updated with the new multi-region depletion capability added to Griffin at the end of FY25.

## Design

Replace the existing multi-region depletion tests with tests using the new multi-region depletion capability merged into Griffin `devel`.

## Impact

Updated tests which actually run and demonstrate the correct syntax for users performing multi-region depletion with the current version of Griffin including executables.

olinwc:
Closes #756

This PR replaces the original implementation of multi-region depletion in Griffin which existed under the `experimental` branch. New capability was merged into Griffin `devel` supporting multi-region depletion at the end of FY25 which is what the modified tests now use.

The functionality is improved since the original implementation in `experimental` required the specification of a "dummy" isotope and XS in a material while the actual calculation of interest was performed in `BatemanVPP`.

This functionality is improved in that the evaluation can be entirely performed in the `MSRDepletionNeutronicsMaterial`. This required adding some transport-related XS data in the microscopic XS file, but this does not change the solution since constant fluxes are specified for the depletion. The transport-related XSs are only required because they are internally expected even if they are not used for a transport solve.

The out-of-core number densities (previously called "off gas" which didn't really make sense since iodine is not a gas) match those previously provided in the existing gold files. Regolding was still necessary because the out-of-core gold files originally tracked every single isotope present, which is no longer the case because of how the out-of-core region WorkingDTLs are constructed.

I also added checking the material densities of the in-core region, because those are really the more important number densities to calculate accurately when performing multi-region depletion.

### DIFF
--- a/doc/content/msr/generic_msr/depletion/model.md
+++ b/doc/content/msr/generic_msr/depletion/model.md
@@ -1,14 +1,17 @@
 # MSR Depletion Model
 
-*Contact: Samuel Walker, Samuel.Walker@inl.gov*
+*Contact: Olin Calvin Olin.Calvin@inl.gov*
 
-*Model link: [MSR Depletion Model](https://github.com/idaholab/virtual_test_bed/tree/devel/msr/generic_msr/depletion)*
+*Model link: [MSR Depletion Model](https://github.com/idaholab/virtual_test_bed/tree/main/msr/generic_msr/depletion)*
 
 The MSR depletion model is an implementation and verification of Griffin's isotope removal capability for two multi-region MSR depletion cases.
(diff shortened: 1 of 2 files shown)


GiudGiud:
Thanks for the patch!

PR #792, opened 2026-04-22
Submodule update

gambka:
This PR was generated by running the following command from the repository root:
python submodule_updater.py ../

Updates application submodules:

Updating submodules
  - apps/bison from 256d088 to origin/master
  - apps/griffin from f42ffc2 to origin/master
  - apps/pronghorn from 0b9c18c to origin/master
  - apps/sockeye from aa8c59b to origin/master
  - apps/dire_wolf from 294b109 to origin/master
  - apps/blue_crab from c4f5073 to origin/master
  - apps/cardinal from cc8f2e4 to origin/master
  - apps/relap-7 from 11b7881 to origin/master
  - apps/moose from e5e72a5 to origin/master
  - apps/ci/msd-tc from 9b8e553 to origin/master
  - apps/mastodon from e749bdd to origin/master
  - apps/grizzly from 93cd00c to origin/master

PR #793, opened 2026-04-22
Updated Sockeye and DireWolf

joshuahansel:
Some tests needed modification due to condenser pool PP being computed on INITIAL now.

Refs #0


gambka:
Lets get this in first and I'll update my PR on the submodule updates here: #792.

gambka (approved):

GiudGiud (changes requested):
DW devel fails an assert in the new code
https://civet.inl.gov/job/3767002/

joshuahansel:
Darn, updated input file to avoid this. Meanwhile I'll make the fix in MOOSE to avoid division by zero, but that won't be needed here.

GiudGiud (approved):

PR #805, opened 2026-05-11
Changed test output for upcoming Sockeye change

joshuahansel:
Sockeye will add another PP to its LCVF model: https://github.inl.gov/ncrc/sockeye/pull/905. A CSV test would diff with the new Sockeye version due to lacking the new PP in the gold file, so we restrict the output to only quantities of interest to avoid this.


gambka (approved):

GiudGiud (approved):
we'll need to reintroduce some PPing later

PR #809, opened 2026-05-22
Update inputs for removal of wall_time_checkpoint parameter

### LINKED ISSUE #808: Address removal of wall_time_checkpoint parameter
## Bug Description

A change at the MOOSE level removing the `wall_time_checkpoint` input parameter is causing some inputs to not run due to unused parameters.

## Steps to Reproduce

Check BISON VTB testing with the latest BISON version.

## Impact

Part of the cause preventing BISON devel to master merges.

gambka:
Closes #808.


GiudGiud (approved):

gambka:
There is still another BISON failure on VTB not related to this that are quite significant diffs. It's coming from a MOOSE submodule update in the last couple days. I don't know the cause of that failure yet.

gambka:
@zachmprince do you suspect that the changes to LHS in https://github.com/idaholab/moose/pull/32833 could be cause of the failure of small_sample parameter study located at htgr/gpbr200/pebble_surrogate_modeling? 

I haven't performed a bisect but this test started to fail about 2 days ago on bison devel testing when attempting to update the MOOSE submodule and that MOOSE PR seems to be the closest thing I could find that would impact this test. The diffs are significant. The larger medium sample test with 1000 samples results remain unchanged though.

zachmprince:
@gambka Yes, that is definitely causing the diffs. If you're willing, you can just regold everything that is showing diffs.

gambka:
> @gambka Yes, that is definitely causing the diffs. If you're willing, you can just regold everything that is showing diffs.

I'll regold if that is the expected change.

zachmprince:
Yes, the changes are expected. The medium and large sampling don't do any file comparison. Unfortunately, the gold file for the small_sample is used in several other tests in the model which will diff. I'll let you decide on whether to also regold those, or keep around the old CSV file so they don't need to regold.

gambka:
@GiudGiud This should be good to go now. Even BC is passing with the submodule update.

GiudGiud (approved):

GiudGiud:
nice thanks for taking care of this

PR #810, opened 2026-05-26
Deprecated test fixes

gambka:
Refs #6.


htgr/assembly/openmc.i
```diff
@@ -287,7 +287,7 @@ num_layers_for_THM = 50 # number of elements in the THM model; for the converged
     from_multi_app = bison
   []
   [heat_flux_to_openmc]
-    type = MultiAppGeneralFieldNearestNodeTransfer
+    type = MultiAppGeneralFieldNearestLocationTransfer
```
GiudGiud, line 290: this is likely going to require at least tuning of the transfers options and possibly debugging of one or both transfers

not impossible just FYI I'm expecting this to be a big task

GiudGiud (commented):
the mechanics ones are straight forward. the transfers I'm not so sure

htgr/assembly/openmc.i
gambka: I've put it as a draft just to get the testing going. I do have the recipes in as scheduled on the weekend to capture further regressions. There are several related to changes in the NavierStokesFV system to using Physics/ approach but the link in the box on the page here: http://127.0.0.1:8000/syntax/Modules/NavierStokesFV/index.html doesn't work for the transition guide. I did find the guide at the bottom of the page and will push an appropriate change to MOOSE to update the broken link.

PR #812, opened 2026-05-27
Update cardinal and cardinal inputs

loganharbour:
## Reason
Update cardinal and inputs that use cardinal.

## Design
- Update cardinal submodule
- Update deprecated parameters in cardinal inputs
- Set memory requirements so that we can limit memory usage for cardinal tests
- Update cardinal tests to use capabilities now that cardinal has a newer moose

## Impact
Updates models using cardinal.

### DIFF
--- a/htgr/assembly/hpc_tests
+++ b/htgr/assembly/hpc_tests
@@ -1,7 +1,7 @@
 [Tests]
   [generate_mesh]
     type = RunApp
-    executable_pattern = 'cardinal*'
+    capabilities = cardinalapp
     input = 'common_input.i solid.i'
     cli_args = '--mesh-only'
     working_directory = 'meshes'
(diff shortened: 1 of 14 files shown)


loganharbour:
Requires a submodule update from https://github.com/neams-th-coe/cardinal/pull/1328 first

GiudGiud (approved):

GiudGiud:
devel recipes need to look at the testroot I ll push a recipe change

loganharbour:
> devel recipes need to look at the testroot I ll push a recipe change

What do you mean? Cardinal doesn't even have debug testing at the moment so I suspect that's why we don't have devel here

GiudGiud:
See the failing recipes at unknown capabilities

Issue #817, opened 2026-05-28
Patch due to merging of meshGenerators in SCM


kyriv-lab:
## Reason
This SCM PR: https://github.com/idaholab/moose/pull/32847
necessitates the editing of the input files in the SCM tests in VTB.

## Design
Edit input files to use the new meshGenerators.

## Impact
Resolve VTB test failures.

PR #818, opened 2026-05-28
Edit input files due to merging of meshgenerators 

kyriv-lab:
This Pr resolves issue #817


GiudGiud:
devel failures unrelated.
We'll have to watch the 7 assemblies case, seems it can diff

PR #822, opened 2026-06-06
Address some HPC bluecrab test failures

GiudGiud:


GiudGiud (commented):
some fixes, though not as many as I would  have liked. 

failures are not due to this PR

PR #827, opened 2026-06-17
Add HTR PM core multiphysics sims

GiudGiud:


GiudGiud:
The devel failures are unrelated. 

I ll likely move thus testing to HPC in the near future.

GiudGiud:
Note we raised an issue in Griffin for achieving a better null/restart with a better decay heat renormalization. 
Feel free to contact us about it if you use the model

Issue #830, opened 2026-06-25
K-HPMR Thermomechanical Model Implementation


soonklee:
## Model Description
The K-HPMR multiphysics model was previously established with thermal-only model encompassing both steady-state and transient analyses. A thermomechanical modeling capability is introduced to the K-HPMR multiphysics model to evaluate structural deformation of the reactor core, assess its feedback effects on neutronics, and provide boundary conditions for future TRISO fuel performance analyses. This thermomechanical extension supports both steady-state operation and load-following transient scenarios.

## Project description
ANL team funded by NEAMS.

## Impact
Example of whole core steady-state and transient multiphysics analysis with displaced mesh.
Improved K-HPMR model by calculating stress propagation.

PR #831, opened 2026-06-25
Add K-HPMR Thermomechanics Model

soonklee:


soonklee:
ref #830

apps/bison
GiudGiud, line 1: let's revert this update, we should not do this with the PR

doc/content/media/htgr/open-xe100/power_comparison.png
GiudGiud, line 1: why is this being updated?

doc/content/media/mrad/K-HPMR_SS_Disp.png
GiudGiud, line 1: these figures need not be that high resolution. 
Please delete them from the commit history then re-add them at <500kb

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [disp_z]
+  [] 
+[]
+
+[Kernels]
+  [null]
+    type = NullKernel
+    variable = temp
+    block = ${hp_blocks}
+  []
```
GiudGiud, line 75:
```suggestion
```
you don't need to do that, just block restrict the variable

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
+# K-HPMR Multiphysics Neutronics-Thermomechanics-Heat Pipe Model
+
+*Contact: Soon Kyu Lee (soon.lee.at.anl.gov), Yinbin Miao (ymiao.at.anl.gov)*
+
+*Model link: [HPMR Model](https://github.com/idaholab/virtual_test_bed/tree/devel/microreactors/mrad)*
```
GiudGiud:
```suggestion
*Model link: [HPMR Model](https://github.com/idaholab/virtual_test_bed/tree/main/microreactors/mrad)*
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
+           codes_used:BlueCrab;Griffin;BISON;Sockeye
+           computing_needs:HPC
+           fiscal_year:2026
+           sponsor:NEAMS
+           institution:ANL
+
+## Inheritance from K-HPMR Model
+
+!alert note title=Acknowledgement
+This model is a thermomechanical extension of the K-HPMR multiphysics model. For the majority of the Multiphysics model description, please refer to the [K-HPMR Multiphysics Model](mrad/mrad_model.md).
```
GiudGiud:
```suggestion
This model is a thermomechanical extension of the K-HPMR multiphysics model. Please refer to the [K-HPMR Multiphysics Model](mrad/mrad_model.md) for additional description of the model.
```

apps/bison
GiudGiud: we used to do this, but now submodule updates are too much work to do on PRs

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
+
+!listing /microreactors/mrad/steady_K_SM/K-HPMR_BISON.i block=Physics/SolidMechanics/QuasiStatic language=text max-height=200px
+
+#### Displacement Diffusion for Outer and Heat Pipe Blocks
+
+The outer blocks - consisting of the outer BeO reflectors, B4C poison blocks, central gas gap blocks, and outer shield — as well as the heat pipe placeholder blocks, are not assigned solid mechanics models to reduce computational cost, as regions of interest are near the fuel compacts. However, the displacement variables must remain defined and seamlessly connected across the entire mesh to avoid numerical artifacts at block interfaces, ensuring convergence in neutronic calculations. To achieve seamless mesh continuity, a `MatDiffusion` kernel with an isotropic diffusivity is applied to each of the three displacement components across the outer and heat pipe block regions. This smoothly propagates the displacement solution from the mechanically active inner assembly into the surrounding regions without introducing artificial mechanical stiffness, ensuring continuity of the displacement field across the full domain.
```
GiudGiud: have you checked if the results are sensitive to the outer core displacement boundary condition? Currently you have 0 "displacement diffusive flux". What about pinning the displacements to 0?

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
+# Multiphysics Results
+
+A comprehensive multiphysics analysis of the 1/6 core HP-MR model was carried out for 
+nominal steady-state operation and a load following transient scenario, coupling the 
+Griffin CFEM-SN(1,3) neutronics solver, the BISON thermomechanical physics solver, and the Sockeye heat pipe solver.
```
GiudGiud:
```suggestion
Griffin DFEM-SN(1,3) neutronics solver, the BISON thermomechanical physics solver, and the Sockeye heat pipe solver.
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
...
+
+These results confirm that, under nominal steady-state conditions, the inclusion of thermal expansion effects introduces negligible perturbations to the neutronics and thermal response of the HP-MR. The effective core geometry change due to thermal deformation is sufficiently small that neither neutron leakage nor the spatial power distribution is meaningfully altered, and the thermal feedback remains essentially unchanged between the two models.
+
+The primary benefit of the thermomechanical model is providing spatially resolved stress and strain distributions throughout the fuel compact ([hpmr_ss_disp]), which are entirely inaccessible from the thermal-only analysis. This enables more rigorous evaluation of TRISO fuel particle performance under irradiation conditions.
```
GiudGiud:
```suggestion
The primary benefit of the thermomechanical model is providing spatially resolved stress and strain distributions throughout the fuel compact ([hpmr_ss_disp]), which are not computed from the thermal-only analysis. This enables more rigorous evaluation of TRISO fuel particle performance under irradiation conditions.
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
...
+!media media/mrad/K-HPMR_SS_Disp.png
+       style=display: block;margin-left:auto;margin-right:auto;width:60%;
+       id=hpmr_ss_disp
+       caption=Steady-state fuel compact and whole core displacements
+
+## Load Following Transient Results
+
+A sudden reduction in secondary-loop heat removal initiates the load following transient. With condenser-side heat removal rendered nearly ineffective, heat pipe temperatures increase, driving up core temperature. This activates the reactor's inherent negative temperature reactivity feedback mechanism that autonomously suppresses reactor power.
+
+This is clearly reflected in the core's thermal and power histories ([hpmr_lf_powtemp]). As the transient begins, average fuel and moderator temperatures increase, reaching a peak of ~900 K near t = 100 s. Simultaneously, the temperature-induced negative reactivity feedback rapidly curtails reactor power, which settles at ~50 kW within 200 seconds with no further significant variation. As this power suppression takes hold, temperatures recede and approach a new equilibrium of 846 K for the fuel and the moderator within the first 1000 seconds of the event.
```
GiudGiud:
```suggestion
This is clearly reflected in the core's thermal and power histories ([hpmr_lf_powtemp]). As the transient begins, average fuel and moderator temperatures increase, reaching a peak of ~900 K near t = 100 s. Simultaneously, the temperature-induced negative reactivity feedback rapidly curtails reactor power, which settles at ~50 kW within 200 seconds with no further significant variation. As this power suppression takes hold, temperatures recede and approach a new equilibrium of 846 K for the fuel and the moderator within the first 1,000 seconds of the event. 
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
...
+
+!media media/mrad/K-HPMR_TR_PowTemp.png
+       style=display: block;margin-left:auto;margin-right:auto;width:100%;
+       id=hpmr_lf_powtemp
+       caption=Time evolution of reactor power, peak, and average fuel compact and moderator temperatures during the load following transient
+
+The structural deformation of core tracks this thermal evolution ([hpmr_lf_disp_mag]). During the initial temperature excursion, thermal expansion causes an immediate increase in core displacement, peaking at a magnitude of approximately $1.4\text{ mm}$. Because the bottom surface is constrained, the largest displacements occur near the upper region of the core. As temperatures stabilize, the thermal expansion relaxes into a new nominal deformation profile.
```
GiudGiud:
```suggestion
The structural deformation of the core tracks this thermal evolution ([hpmr_lf_disp_mag]). During the initial temperature excursion, thermal expansion causes an immediate increase in core displacement, peaking at a magnitude of approximately $1.4\text{ mm}$. Because the bottom surface is constrained, the largest displacements occur near the upper region of the core. As temperatures stabilize, the thermal expansion relaxes into a new nominal deformation profile.
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
...
+
+!media media/mrad/K-HPMR_TR_Mag.png
+       style=display: block;margin-left:auto;margin-right:auto;width:50%;
+       id=hpmr_lf_disp_mag
+       caption=Average displacement as a function of core height and time during the load following transient.
+
+Examination of the displacement components ([hpmr_lf_disp_comp]) indicates that the overall deformation is dominated by axial expansion (disp_z). The resulting deformation field remains continuous throughout the transient, indicating that the load following event produces a predictable, thermally driven expansion rather than introducing localized mechanical distortion or significant differential expansion.
```
GiudGiud: > rather than introducing localized mechanical distortion or significant differential expansion.

you're not modeling anything that could do that though. If you were modeling contact with some form of grid/plate holding the assemblies you could maybe see strain concentrations

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [heat_conduction]
```
GiudGiud, line 76: this is using a quasi-static model forheat conduction

did I miss this in the documentation?

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [diff_z]
+    type = MatDiffusion
+    block = '${outer_blocks} ${hp_blocks}'
+    variable = disp_z
+    diffusivity = 1e5
+    extra_vector_tags = 'ref'
+  []
+[]
+
+[Physics/SolidMechanics/QuasiStatic]
```
GiudGiud, line 113: should also mention in the doco that this is doing quasi-static not dynamic

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+    normal_factor = '${fparse corr_factor}'
+  []
+[]
+
+[BCs]
+  [outside_bc]
+    type = ConvectiveFluxFunction # (Robin BC)
+    variable = temp
+    boundary = 'bottom top side'
+    coefficient = 1e3 # W/K/m^2
```
GiudGiud, line 281: isnt this high for air?

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+    T_infinity = 800 # K air temperature at the top of the core
+  []
+  [hp_temp]
+    type = CoupledConvectiveHeatFluxBC
+    boundary = 'heat_pipe_ht_surf'
+    variable = temp
+    T_infinity = hp_temp_aux
+    htc = 750
```
GiudGiud, line 289: @joshuahansel thoughts on this value for HP <-> medium

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [no_y]
+    type = DirichletBC
+    variable = disp_y
+    boundary = 'centerline'
+    value = 0.0
+  []
+  [no_z]
+    type = DirichletBC
+    variable = disp_z
+    boundary = 3000
```
GiudGiud: please use boundary names

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [disp_y_trans]
+  []
+[]
+
+[AuxKernels]
+  [assign_disp_x_trans]
+    type = NormalizationAux
+    variable = disp_x_trans
+    source_variable = disp_x
+    execute_on = 'NONLINEAR'
```
GiudGiud: why  NONLINEAR? isnt this to be transfered? It's not needed for the nonlinear solve

GiudGiud: is it to stagger it with the correction one?
If so mention it, we will have auxkernels execution groups at some point this summer which will make this unnecessary

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+    thermal_conductivity_scale_factor = 1.0
+    fast_neutron_fluence = 0
+    temperature = temp
+  []
+  [moderator_thermal]
+    type = HeatConductionMaterial
+    block = '${yh_blocks}'
+    temperature = temp
+    thermal_conductivity = 20 # W/m/K
+    specific_heat = 500 # random value
```
GiudGiud: isnt the moderator thermal response fairly important ? Are there no values in the litterature?

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+    temperature = temp
+    block = '${outer_ref_blocks}'
+    fluence_conversion_factor = 1.0
+  []
+  [airgap_thermal]
+    type = HeatConductionMaterial
+    block = ${air_blocks} # Helium gap
+    temperature = temp
+    thermal_conductivity = 0.15 # W/m/K
+    specific_heat = 5197 # random value
```
GiudGiud: no this seems about right but give a reference

GiudGiud: maybe a section in the doco if you don't want to add comments everywhere in the Materials section of both inputs

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,724 @@
...
+  [from_sockeye_temp]
+   type = MultiAppGeneralFieldNearestLocationTransfer
+   from_multi_app = sockeye
+   source_variable = hp_temp_aux
+   variable = hp_temp_aux
+   execute_on = 'timestep_begin'
+ []
+ [to_sockeye_flux]
+   type = MultiAppGeneralFieldNearestLocationTransfer
+   variable = master_flux
```
GiudGiud:
```suggestion
   variable = total_flux
```
let's rename the variable

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
```diff
@@ -0,0 +1,360 @@
...
+    zmin = -0.0
+    zmax = 2.1
+  []
+  [assign_coarse_id]
+    type = CoarseMeshExtraElementIDGenerator
+    input = fmg_id
+    coarse_mesh = coarse_mesh
+    extra_element_id_name = coarse_element_id
+  []
+ uniform_refine = 0
```
GiudGiud:
```suggestion
  uniform_refine = 0
```

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
```diff
@@ -0,0 +1,360 @@
...
+  []
+[]
+
+[TransportSystems]
+  particle = neutron
+  equation_type = transient
+
+  G = 11
+  VacuumBoundary = '10000 2000 3000'
+  ReflectingBoundary = '147'
```
GiudGiud: could you use names instead?

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,449 @@
+################################################################################
```
GiudGiud, line 1: @joshuahansel if you want to review this sockeye input

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,449 @@
...
+#    function = 1.0
+#    execute_on = 'INITIAL TIMESTEP_END'
+#  []
+  [operational_pp]
+    type = ElementAverageValue
+    variable = operational_aux
+    execute_on = 'initial timestep_begin TIMESTEP_END'
+  []
+  # set `catastrophic_pp` as if it is recoverable for a solving to steady-state simulation
+  # this MUST be changed back to catestrophic in transient simulations
```
GiudGiud:
```suggestion
  # this MUST be changed back to catastrophic in transient simulations
```

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,449 @@
...
+  [catastrophic_pp]
+    type = HeatRemovalRateLimitScale
+    heat_addition_pps = 'evaporator_boundary_integral'
+    limit_condenser_side = false
+    recoverable_heat_removal_limit_pps = 'hp_boiling_limit hp_capillary_limit hp_entrainment_limit'
+    catastrophic_heat_removal_limit_pps = ''
```
GiudGiud: this is the transient input and the advice above says to flip it to catastrophic

microreactors/mrad/mesh/gold/HPMR_OneSixth_Core_meshgenerator_tri_hp_fine.e
GiudGiud, line 1: please use LFS for this

GiudGiud: (you will need to delete it from the commit history first)

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
```diff
@@ -0,0 +1,360 @@
...
+  verbose = true
+
+  richardson_abs_tol = 5e-4
+  richardson_rel_tol = 5e-4
+
+  richardson_max_its = 100
+  inner_solve_type = GMRes
+  max_inner_its = 100
+
+  fixed_point_max_its = 1
```
GiudGiud, line 86: so are we running 1 bison simulation per griffin step?
or is this being overridden by the richardson parameter

GiudGiud (changes requested):
is the technical report out for this model? I don't see any new reference. 

The approach for mechanics seems similar to Kendrick's thesis at MIT, you might want to cite it there

It would be nice to point to mesh convergence studies in that report too.

Have we done homogenized thermo-mechanics at the core scale before @gambka ? Is it an established methodology?

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
```diff
@@ -0,0 +1,360 @@
...
+  [Tf]
+    # initial_condition = 873.15
+    order = CONSTANT
+    family = MONOMIAL
+  []
+  [nH]
+    # initial_condition = 1.94
+    order = CONSTANT
+    family = MONOMIAL
+  []
```
GiudGiud: this input seems to be geared up for hydrogen migration studies, but I dont see any reference for it in the doco or any solve in the thermal solver?

doc/content/media/mrad/K-HPMR_SS_Disp.png
soonklee: Figure sizes are reduced to <500kb.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
soonklee: revised to main.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
soonklee: Revised as suggested.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: Revised to DFEM

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: Revised as suggested.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: Comma added.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: Revised.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: Paragraph is revised to clearly describe model behavior and its limitations.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
soonklee: revised.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
soonklee: We are running 1 BISON solve per Griffin timestep. I think richardson iteration is internal to neutronics solve.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
soonklee: revised to sideset names.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
soonklee: nH related variable/postprocessors are removed. We shared initial input.

microreactors/mrad/mesh/gold/HPMR_OneSixth_Core_meshgenerator_tri_hp_fine.e
soonklee: LFS pointer added and cleaned commit history.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: block restriction made on variable

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: We are using quasistatic model. Model description is revised to mention quasi-static.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: quasi-static mentioned in model description.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: Yes. disp_x/y_trans aren't transferred anywhere, but feed to disp_x/y_trans_corr, which runs at timestep_end and needs disp_x/y_trans to hold the converged value from the last nonlinear iteration. description added in Transfer of Displacement Data to Griffin section.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: airblock represents helium. technical reports are referenced for details in thermal models.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: revised to boundary names.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: technical reports are referenced for details in thermal models.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: Technical reports are referenced for details in thermal models.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: variable renamed to total_flux

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
soonklee: revised.

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
soonklee: comment removed.

gambka:
> is the technical report out for this model? I don't see any new reference.
> 
> The approach for mechanics seems similar to Kendrick's thesis at MIT, you might want to cite it there
> 
> It would be nice to point to mesh convergence studies in that report too.
> 
> Have we done homogenized thermo-mechanics at the core scale before @gambka ? Is it an established methodology?

We've used homogenization techniques for thermal solutions for other applications. I'm not aware of full core scale thermo-mechanics applications, but I'd have to read up on the approach used here. There are established homogenization methods. Applications of these techniques have been mostly applied to upscaling of lower-length scale models and their information into engineering scale studies. An example is provided here: [10.1016/j.commatsci.2014.12.039](https://doi.org/10.1016/j.commatsci.2014.12.039)

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
soonklee: The sensitivity to the outer core displacement boundary condition has not been checked, but zero-flux is a deliberate choice rather than a default. As none of the outer blocks carry eigenstrain or elasticity tensor, pinning the displacement to zero at the outer boundary would suppress the expansion these regions would otherwise undergo, introducing constraint. The zero-flux condition instead allows the diffused field to passively follow the trend set by the mechanically active inner assembly.

soonklee (commented):
Thanks for the thorough review @GiudGiud. Comments are addressed below.

Changes include:
- Figures resized to <500kb as requested
- Mesh file properly tracked with git lfs, including cleaning the raw binary out of early commit history
- Markdowns revised as requested
- Input files revised as requested, plus a couple of minor clarifying comments added

The model is updated and ready for further review.
Note that this branch was rebased to fix the lfs history.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
GiudGiud: have you checked that this is enough ? Not converging the multiphysics coupling could change results

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
+
+!tag name=MRAD Micro-Reactor Multiphysics Thermomechanical model
+     description=A core multiphysics model coupling neutronics, thermomechanics, and heat pipe thermal performance for steady state and transient simulations
+     image=https://mooseframework.inl.gov/virtual_test_bed/media/mrad/hpmr_mesh_proc.png
+     pairs=reactor_type:microreactor
+           reactor:HPMR
+           geometry:core
+           simulation_type:multiphysics
+           V_and_V:demonstration
+           input_features:multiapps;reactor_meshing;mixed_restart;solid_mechanics
```
GiudGiud:
```suggestion
           input_features:multiapps;reactor_meshing;mixed_restart
```
let's not add new features that were not added to the other inputs

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
...mechanics-heat pipe model extends the thermal-only multiphysics model described in the [K-HPMR Multiphysics Model](mrad/mrad_model.md) by incorporating solid mechanics into the BISON application. The HPMR model was originally developed in FY21 [!citep](Stauff2021Detailed), and the thermal-only multiphysics model referenced above was further developed, with a preliminary thermomechanical model implemented in FY24 [!citep](Miao2025) using simplified stress boundary conditions on the symmetry planes; see these references for additional details on the underlying thermal models. This treatment is enhanced here to correctly impose mirror pressure boundary conditions, as described in the Stress Boundary Conditions section below. Thermally induced deformations of the reactor core materials are computed by BISON alongside the thermal solution, and the resulting displacement fields are transferred to Griffin to update the neutron transport calculation on the displaced mesh. This coupling allows the model to account for dimensional deformations in the fuel compacts, moderator, monolith, and reflector materials — and their consequent effect on reactivity — under realistic operating conditions.
```
GiudGiud:
```suggestion
The neutronics-thermomechanics-heat pipe model extends the multiphysics (neutronics-thermal-heat pipe) model described in the [K-HPMR Multiphysics Model](mrad/mrad_model.md) by incorporating solid mechanics into the BISON application (previously only solving for temperature). The HPMR model was originally developed in FY21 [!citep](Stauff2021Detailed), and the thermal-only multiphysics model referenced above was further developed, with a preliminary thermomechanical model implemented in FY24 [!citep](Miao2025) using simplified stress boundary conditions on the symmetry planes; see these references for additional details on the underlying thermal models. This treatment is enhanced here to correctly impose mirror (reflective) pressure boundary conditions, as described in the Stress Boundary Conditions section below. Thermally induced deformations of the reactor core materials are computed by BISON alongside the thermal solution, and the resulting displacement fields are transferred to Griffin to update the neutron transport calculation on the displaced mesh. This coupling allows the model to account for dimensional deformations in the fuel compacts, moderator, monolith, and reflector materials — and their consequent effect on reactivity — under realistic operating conditions.
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
+
+The solid mechanics action is activated for all structurally active inner reactor material blocks: fuel compacts, graphite monolith, YH moderator, stainless steel moderator cladding, and the axial reflectors. The mechanical solution is quasistatic - the momentum balance neglects inertial effects, so the model captueres thermally induced deformation but not dynamic or vibrational response. Material-specific thermal expansion eigenstrains are computed for each inner assembly block. For the fuel compact blocks, the `GraphiteMatrixThermalExpansionEigenstrain` model is used to capture the thermal expansion behavior of the TRISO-loaded graphite matrix. For the graphite monolith blocks, the `GraphiteGradeThermalExpansionEigenstrain` model is applied using grade G-348 properties, reflecting the distinct graphite grade used for the structural monolith relative to the fuel matrix. For the YH moderator blocks, a constant isotropic thermal expansion coefficient is assigned on the `ComputeThermalExpansionEigenstrain` model. The stainless steel moderator cladding uses the `SS316ThermalExpansionEigenstrain`, and the inner axial reflectors use the `BeOThermalExpansionEigenstrain` model.
```
GiudGiud:
```suggestion
The solid mechanics action is activated for all structurally active inner reactor material blocks: fuel compacts, graphite monolith, YH moderator, stainless steel moderator cladding, and the axial reflectors. The mechanical solution is quasistatic - the momentum balance neglects inertial effects, so the model captures thermally induced deformation but not dynamic or vibrational response. Material-specific thermal expansion eigenstrains are computed for each inner assembly block. For the fuel compact blocks, the `GraphiteMatrixThermalExpansionEigenstrain` model is used to capture the thermal expansion behavior of the TRISO-loaded graphite matrix. For the graphite monolith blocks, the `GraphiteGradeThermalExpansionEigenstrain` model is applied using grade G-348 properties, reflecting the distinct graphite grade used for the structural monolith relative to the fuel matrix. For the YH moderator blocks, a constant isotropic thermal expansion coefficient is assigned on the `ComputeThermalExpansionEigenstrain` model. The stainless steel moderator cladding uses the `SS316ThermalExpansionEigenstrain`, and the inner axial reflectors use the `BeOThermalExpansionEigenstrain` model.
```

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
GiudGiud: but why would these outer regions expand at that rate? They are not heated to the same temperature.
Besides wouldn't the shield be held externally?

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
GiudGiud: sounds good thanks for letting me know. I can update this once the improvement in MOOSE has been done

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Steady-State Case Simulation
+
+The steady-state simulation establishes the nominal operating condition of the K-HPMR thermomechanical model, like the counterpart simulation of the thermal-only model, including the thermally deformed core geometry that results from full-power operation. This solution also provides the initial condition for subsequent transient analyses.
+
+## Load Following Transient Simulation
+
+The load following transient is driven by a sharp degradation in the secondary-loop heat removal system, implemented by modifying the condenser envelope boundary condition in the Sockeye model, reducing the convective heat transfer coefficient by 99.99%, thereby eliminating heat extraction from the secondary loop. The resulting rise in condenser temperature propagates through the coupled BISON thermomechanical model, elevating the core temperatures and reducing reactor power through negative reactivity feedback in Griffin. Beyond what was captured in thermal-only model, the thermomechanical model additionally tracks the evolution of core mechanical deformation as local temperature field fluctuates throughout the transient.
```
GiudGiud: why call this a load follow for this heat pipe transient? cutting the secondary seems more like an equivalent to ULOFC than a load follow?
@miaoyinb might want to pitch in too

microreactors/mrad/steady_K_SM/K-HPMR_SOCKEYE.i
```diff
@@ -0,0 +1,446 @@
...
+[Executioner]
+  type = Transient
+
+  solve_type = NEWTON
+  petsc_options_iname = '-pc_type'
+  petsc_options_value = 'lu'
+  scheme = bdf2
+  line_search = none
+
+  # ensure nl_abs_tol >> nl_rel_tol
```
GiudGiud: you don't do this below though

also this is bad advice imo

microreactors/mrad/steady_K_SM/K-HPMR_GRIFFIN.i
```diff
@@ -0,0 +1,282 @@
...
+  plus = true
+  dbgmat = false
+  grid_names = 'Tfuel'
+  grid_variables = 'Tf'
+[]
+
+[Materials]
+  [mod]
+    type = CoupledFeedbackMatIDNeutronicsMaterial
+    block = '200 203 100 103 301 303 10 503 600 601 201 101 400 401 250'
```
GiudGiud: could you use mod_blocks instead here?

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
soonklee: Quick sensitivity study was ran increasing fixed_point_max_its from 1 to 5. The difference in fuel and moderator temperatures between the two cases was less than 0.01 K throughout the transient, confirming a single fixed-point iteration per Griffin timestep is sufficient for this case.

soonklee (commented):
Minor revisions made to inputs.

microreactors/mrad/steady_K_SM/K-HPMR_GRIFFIN.i
soonklee: block ids replaced with block names.

microreactors/mrad/steady_K_SM/K-HPMR_SOCKEYE.i
soonklee: comment removed.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_model.md
soonklee: terminology follows the original mrad multiphysics analyses (Stauff et al. 2022, 2023), where the scenario is referred to as a load following transient, where the intent is to demonstrate the reactor's inherent passive load following capability when secondary side heat removal system is removed. ULOFC may imply an accident, whereas here the reactor response through its inherent feedback is without any accident classification or safety system actuation.

soonklee: References were added to help readers.

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
joshuahansel: Sorry, I never got an email for this for some reason. I'll take a look.

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,446 @@
...
+    make_pressure_corrections = true
+  []
+
+  [condenser_boundary]
+    type = HSBoundaryAmbientConvection
+    boundary = 'hp:cond:outer'
+    hs = hp
+    T_ambient = ${T_ext_cond}
+    htc_ambient = ${htc_ext_cond} #large value to approach an effective DirichletBC
+#    scale_pp = bc_scale_pp
```
joshuahansel: It seems that you are not doing any sort of scaling, so I expect you effectively have just a heat pipe that is not limited at all. You should either delete all of the limit-related PPs and such and acknowledge those are your assumptions, or you should repair it. Note that the approach taken here seems to be outdated. What we typically do now is we do not scale the BCs; we scale the core thermal conductivity with `[ControlLogic]`. There is an example in `sockeye/examples/conduction_model/variable_conductivity/variable_conductivity.i`. I can help with any questions if you go that route.

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,446 @@
...
+    type = PotassiumTwoPhaseFluidProperties
+    emit_on_nan = none
+  []
+[]
+
+[Components]
+  [hp]
+    type = HeatPipeConduction
+
+    # Common to both HeatPipe2Phase and HeatPipeBlackbox
```
joshuahansel: These distinctions are not important in this input; remove associated comments

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
```diff
@@ -0,0 +1,69 @@
+# Multiphysics Results
```
joshuahansel, line 1: Please add an HPC resources table as shown on https://virtualtestbed.inl.gov/template.html to tell users what kind of resources were used to run your models.

joshuahansel (changes requested):

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
soonklee: Thank you for the suggestion. the operational limit pps and the scaling have been removed from the model to avoid the misleading implication that limits are being enforced. We'll plan to implement the control logic block referencing the example in a future update.

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
soonklee: comments are removed.

doc/content/microreactors/mrad/mrad_k_hpmr_thermomechanics_results.md
soonklee: run command and hpc resources are added.

joshuahansel (approved):

GiudGiud:
Needs testing. If you're not adding it I'll do it when I have time

soonklee:
> Needs testing. If you're not adding it I'll do it when I have time

I haven't written a tests spec before, but I'll get one set up and push it to PR soon.

soonklee:
I added the initial regression test structure following the existing examples. I would appreciate feedback on whether the current test settings are appropriate.

microreactors/mrad/mesh/tests
```diff
@@ -27,4 +27,13 @@
         cli_args = "--mesh-only 'HPMR_OneSixth_Core_meshgenerator_tri_rotate_bdry_fine.e'"
         recover = false
     []
+    [mesh_generation_bison_hp]
+        type = 'Exodiff'
+        input = 'HPMR_OneSixth_Core_meshgenerator_tri_hp_fine.i'
+        exodiff = 'HPMR_OneSixth_Core_meshgenerator_tri_hp_fine.e'
```
GiudGiud, line 33: did you check in the mesh file?
if so and using LFS, we'll need to modify virtual_test_bed/.civet/griffin.yml (and the others) to download then file AND add a sym link to it in mrad/mesh/gold

microreactors/mrad/steady_K_SM/tests
```diff
@@ -0,0 +1,74 @@
...
+
+  # Regression test
+  [coupled_simulation]
+    type = CSVDiff
+    input = 'K-HPMR_GRIFFIN.i'
+    csvdiff = 'K-HPMR_GRIFFIN_out.csv'
+    executable_pattern = 'dire_wolf*'
+    cli_args = "Executioner/richardson_max_its=2 Executioner/richardson_abs_tol=1e-1 Outputs/csv=true "
+               "Outputs/file_base=K-HPMR_GRIFFIN_out Outputs/exodus=false MultiApps/bison/cli_args='Outputs/exodus=false'"
+    min_parallel = 48
```
GiudGiud: unfortunately too much. create an hpc_tests file and the tests for that model will run there

microreactors/mrad/steady_K_SM/tests
```diff
@@ -0,0 +1,74 @@
...
+    check_input = True
+    executable_pattern = 'sockeye*|dire_wolf*'
+    allow_warnings = true
+  []
+
+  # Regression test
+  [coupled_simulation]
+    type = CSVDiff
+    input = 'K-HPMR_GRIFFIN.i'
+    csvdiff = 'K-HPMR_GRIFFIN_out.csv'
```
GiudGiud, line 32: do you have that file pre-generated? it will need to go in microreactors/mrad/steady_K_SM/gold

microreactors/mrad/mesh/tests
soonklee: Yes, it's tracked via LFS under `microreactors/mrad/mesh/gold/HPMR_OneSixth_Core_meshgenerator_tri_hp_fine.e`, which is already covered by `lfs_include` in `.civet/griffin.yaml`. Confirmed git `check-attr` shows `filter: lfs` and `git lfs ls-files` shows the object is pushed. No `griffin.yaml` changes or symlink needed since it's already in the whitelisted `gold/` path.

microreactors/mrad/steady_K_SM/tests
soonklee: Yes, it is pre-generated and placed in mrad/steady_K_SM/gold.

microreactors/mrad/mesh/tests
GiudGiud: nice

microreactors/mrad/steady_K_SM/tests
soonklee: I'll revise tests and hpc_tests on mrad correctly.

soonklee:
Hi @GiudGiud, 
Please see following updates for tests/hpc_tests. 
Latest push includes:
- Fix and regenerate `hp_fine` mesh input and gold to correctly designate `outer_reflector` block for bison
- Document `NullKernel` rationale for `temp` on `hp_blocks` (Sockeye owns heat pipe physics)
- Regenerate `steady_K_SM` and `load_following_K_SM` gold CSVs
- Add `ss` symlink for `load_following_K_SM`, following the existing `Na` convention
- List standalone tests for Griffin and Sockeye on `mrad/steady_K_SM/tests` and for Bison on `mrad/hpc_tests` due to fine mesh
- Confirmed steady/load_following coupled_tests on `mrad/hpc_tests` with loosen tolerance

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
soonklee: I looked into removing null block, but wanted to explain why I eneded up keeping it rather than block-restricting `temp` from `hp_blocks`. `temp` on `hp_blocks` wasn't meant to be physically meaningful as Sockeye owns the heat pipe thermal-fluid physics, and coupling occurs through `hp_temp_aux` at `heat_pipe_ht_surf` on the solid side. The reason it needs to be included is that `hp_block` still needs to be part of the solid mechanics mesh, as heat pipes have to deform coherently with the rest of the core geometry. The mesh boundary/sideset definitions of `top`, `bottom`, `side_mirror` are whole-geometry constructs spanning the full model including `hp_blocks`. If `temp` doesn't cover `hp_blocks`, every `BCs` and `postprocessor` referencing those shared sidesets breaks with a "variable not defined on this boundary" error. 

I did consider the alternative ways, but those required surgically touching every `BCs` and `postprocessor` referencing those boundaries, with more places for a future edit to silently reintroduce the same failure. While `NullKernel` isn't ideal for a genuine nonlinear variable, since `temp` is a single field shared across both `non_hp_blocks` and `hp_blocks` (bookkeeping only), it can't be split into two variable types within a system. Given that constraint, this seemed like the right tool for the job. Rational for using NullKernel is specified within the input.

soonklee (commented):
Pushed gold and test matrix. Ready for re-review.

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,705 @@
+################################################################################
+## NEAMS Micro-Reactor Application Driver                                     ##
+## Heat Pipe Microreactor Steady State                                        ##
```
GiudGiud:
```suggestion
## Heat Pipe Microreactor load follow                                      ##
```

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
GiudGiud: Excluding selected blocks from a sideset takes a single ParsedGenerateSideset with an "exclude_subdomains" argument

>  The mesh boundary/sideset definitions of top, bottom, side_mirror are whole-geometry constructs spanning the full model including hp_blocks. If temp doesn't cover hp_blocks, every BCs and postprocessor referencing those shared sidesets breaks with a "variable not defined on this boundary" error.

this is exactly the argument for adding specific boundaries for temperature. you're currently applying boundary conditions on sidesets facing elements where only a nullkernel exists. The effect of that is unknown.

microreactors/mrad/load_following_K_SM/K-HPMR_GRIFFIN_tr.i
```diff
@@ -0,0 +1,318 @@
+################################################################################
+## NEAMS Micro-Reactor Application Driver                                     ##
+## Heat Pipe Microreactor Steady State                                        ##
```
GiudGiud:
```suggestion
## Heat Pipe Microreactor Load following transient                         ##
```

microreactors/mrad/load_following_K_SM/K-HPMR_BISON_tr.i
```diff
@@ -0,0 +1,705 @@
...
+## BISON Child Application input file                                         ##
+## Thermomechanical model (Heat conduction, Thermal Expansion, Thermal Stress)##
+## FY26 Summer Update for Solid Mechanics Implementation                      ##
```
GiudGiud, line 6:
```suggestion
## of partially-homogenized geometry                                                              ##
## FY26 Summer Update for Solid Mechanics Implementation                      ##
```

microreactors/mrad/load_following_K_SM/K-HPMR_SOCKEYE_tr.i
```diff
@@ -0,0 +1,398 @@
+################################################################################
+## NEAMS Micro-Reactor Application Driver                                     ##
+## Heat Pipe Microreactor Steady State                                        ##
+## Sockeye Grandchild Application input file                                  ##
```
GiudGiud:
```suggestion
## Heat Pipe Microreactor Load following transient                       ##
## Sockeye Grandchild Application input file                                  ##
```

GiudGiud (changes requested):
testing looks good. I'll turn it on right now. 

I ll send you an invite to be part of the VTB organization on github so the tests run automatically. 
The Direwolf HPC recipe is the one we ll need to watch here

GiudGiud:
the runs are going OOM
you might want to turn on distributed mem and add a hpc_mem_per_cpu (see other tests specs) argument to use less cores per node (but still requesting lots of memory)

soonklee:
> the runs are going OOM you might want to turn on distributed mem and add a hpc_mem_per_cpu (see other tests specs) argument to use less cores per node (but still requesting lots of memory)

Increasing ranks helped a little (resolved syntax check OOM), and currently adjusting hpc_mem_per_cpu to resolve direwolf hpc checks.

soonklee:
Hi @GiudGiud, I wanted to flag where I am at on the mrad hpc_tests for this PR before I push further. [thermomechanics_bison] is passing cleanly now (96 ranks 15G/cpu). [coupled_simulation_steady_sm] (Griffin-Bison-Sockeye) is not quite passing through. I tested various rank/mem combinations via CIVET - OOM kills, MPI launcher fails, and others. This coupled run with solid mechanics require higher resources than what was verified. 

Given [thermomechanics_bison], [thermomechanics_bison_syntax_load], and the Griffin/Sockeye standalones all pass, and the coupling itself is already exercised by the thermal-only model, I'd like to propose removing [coupled_simulation_steady_sm] and [coupled_simulation_load_follow_sm] from this PR rather than continue sinking CIVET time into resource tuning. Please share your thoughts. Thanks!

GiudGiud:
> [coupled_simulation_steady_sm] (Griffin-Bison-Sockeye) is not quite passing through.

what were you using the first time it was ran? like when people were doing analysis with it

soonklee:
> what were you using the first time it was ran? like when people were doing analysis with it

coupled_simulation_steady_sm test was golded at 96 ranks on Sawtooth. But at 96 ranks on the higher-fidelity config, it OOMs; it only runs cleanly at 400–500 ranks. coupled_simulation_steady_sm test runs at 150 ranks/15G but fails with timeout, taking much longer than expected.

GiudGiud:
so let's run it at 500 ranks ?
How long did it take with that many ranks?

microreactors/mrad/hpc_tests
```diff
@@ -131,10 +132,11 @@
...
     cli_args = "Executioner/richardson_max_its=5 Executioner/richardson_abs_tol=1e-1 Outputs/csv=true "
                "Executioner/end_time=5  Outputs/file_base=mrad_coupled_load Outputs/exodus=false "
-               "MultiApps/bison/cli_args='Outputs/exodus=false;Outputs/csv=false;Problem/restart_file_base=../steady/mrad_coupled_out_bison0_cp/LATEST;Mesh/file=../steady/mrad_coupled_out_bison0_cp/LATEST' "
+               "MultiApps/bison/cli_args='Outputs/exodus=false;Outputs/csv=false;Problem/restart_file_base=../steady_K_SM/mrad_coupled_out_bison0_cp/LATEST;Mesh/file=../steady_K_SM/mrad_coupled_out_bison0_cp/LATEST' "
                "--distributed-mesh"
-    min_parallel = 150
-    hpc_mem_per_cpu = 15G
+    min_parallel = 500
+    max_parallel = 500
+    hpc_mem_per_cpu = '23000M'
```
GiudGiud, line 139: 500 ranks is 10 nodes (using sawtooth numbers, dont recall windriver), but at 23GB per CPU it's actually 60 nodes

how much do you actually need? what were the settings of the original run back in analysis?

soonklee: I re-tested on windriver since CIVET was running on windriver, and reserving 500 ranks without extra room resulted in OOM. 23GB is reverse-calculated from the actual manual sbatch config used for the original analysis run. I used following: 50 nodes, 500 rank total, --ntask-per-node=10, --mem=0. That is ~239GiB/node split across 10 ranks/note = 23.9GB/rank. I chose 23G to force slurm into that same 10 rank-per-node density. But, the latest CIVET test resulted in system crash (exit 112). So far I tried both a regular rank count and a manual config copying setup, and neither survives cleanly.

I'd like to renew my earlier proposal to drop coupled_simulation_steady_sm and coupled_simulation_load_follow_sm form hpc_tests. Standalones cover script and thermal-only coupled test covers actual multiapp coupling scheme. I can revisit CIVET inclusion later as we will continue to develop this model. Let me know if that works for you or if there's another angle worth trying first. Thanks.

soonklee: Hi @GiudGiud, just following up on this conversation in case it got buried. I’m happy to remove coupled_simulation_steady_sm and coupled_simulation_load_follow_sm from hpc_tests as proposed, unless you’d prefer that I try another approach first. Thanks!

GiudGiud: hello

not buried I just won't have time to work on this for the VTB until next week!
but you PR is first on the list once we get there

PR #836, opened 2026-07-09
Update BISON submodule to fix BISON devel testing

gambka:
Refs #0.


gambka:
The BISON HPC failure is the same recipe failure we see on scheduled. Not a model failure.

GiudGiud (approved):
bison devel is fixed

Issue #834, opened 2026-07-06
Segregated version of the generic PBR tutorial model


alb-sey:
## Model Description

The segregated PBR model is the steady-state SIMPLE version of the Newton Generic Pebble Bed PBR tutorial. It models the same reactor physics and geometry, but uses a steady-state SIMPLE solver and runs with Pronghorn. It is intended to reach the same steady operating condition more directly than the pseudo-transient Newton version. The model is small enough to run on a workstation; HPC is not required.

## Project description

This model is using the SIMPLE solver for porous-medium thermal hydraulics, developed by Peter German. The input files were modified by Alberic Seydoux as a part of his internship project, in order to evaluate the SIMPLE solver.

## Impact

This model is used to assess the SIMPLE porous-medium capabilities, and provide an improvement over the existing pseudo-transient model running on the Newton solver.

PR #841, opened 2026-07-21
Simple generic pbr

alb-sey:
LRS# : INL/MIS-26-93929 

Issue #834 

The segregated PBR model is the steady-state SIMPLE version of the Newton Generic Pebble Bed PBR tutorial. It models the same reactor physics and geometry, but uses a steady-state SIMPLE solver and runs with Pronghorn.


alb-sey:
The PR contains the input files, documentation, and media associated with the segregated PBR

PR #843, opened 2026-07-30
Update MeshInfo tests

loganharbour:
Updates tests to use MeshInfo in a more appropriate manner.

Resolves failures in https://civet.inl.gov/job/4021756/.

Refs and requires idaholab/moose#33437


loganharbour:
@GiudGiud this is what I have in mind. I've only done one test so far, but here we output bounding boxes and volumes for sidesets and subdomains. Should be a better diff of what exists there. What do you think?

GiudGiud (approved):

GiudGiud:
works for me. we can add more metrics as we get comfortable with what we can control from the meshing script

loganharbour:
I still have another test to fix, and this will require a MOOSE update

GiudGiud (approved):

PR #851, opened 2026-08-10
Add min_slots, loosen tolerances due to compiler optimization flags

loganharbour:
refs https://github.com/idaholab/moose/pull/33466

### DIFF
--- a/htgr/mhtgr/3D_mesh/tests
+++ b/htgr/mhtgr/3D_mesh/tests
@@ -4,6 +4,7 @@
     type = RunApp
     input = 3D_mhtgr_final.i
     check_input = True
+    min_slots = 5 # memory heavy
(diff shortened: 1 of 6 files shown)


GiudGiud (approved):
should we document the slot size you're using for these selections?
if we increase it we could lower the min_slots accordingly

loganharbour:
> should we document the slot size you're using for these selections?

what do you want documented? current usage?

GiudGiud:
basically the memory usage you are limiting when setting min_slots of 5

loganharbour:
> basically the memory usage you are limiting when setting min_slots of 5

I don't have anything explicit yet, I was just dropping these when I saw it. I was targeting < 1GB/slot, but that's not official yet

PR #853, opened 2026-08-16
Regold Griffin HPC tests

NamjaeChoi:


GiudGiud (approved):
nothing to do from this side on this ABTR test, it just won't be consistent

GiudGiud:
thx for the patch!

NamjaeChoi:
Oh actually it should be regolded too. It uses the result of `xs_generation`, so it wasn't run before.

PR #854, opened 2026-08-16
Regold more Griffin HPC test

NamjaeChoi:


GiudGiud (approved):
thanks for the patch!

PR #860, opened 2026-08-24
Avoid testing first step of MR mechanics since we are essentially comparing 0s

GiudGiud:
refs #0


GiudGiud:
bison testing of these models passed.
we might need to expand this to BC later

PR #871, opened 2026-08-27
Update mastodon submodule after container changes

loganharbour:
refs idaholab/moose#33466


GiudGiud (approved):
HPC testing failures look unrelated and I'm not worried there

PR #873, opened 2026-08-28
Update dire_wolf submodule after container changes

loganharbour:
refs idaholab/moose#33466


GiudGiud (approved):
last broken test is fixed in new DW. not worth re-running the suite for it

GiudGiud:
DW HPC has failures, which are not related and to be dealt with in submoduleupdate

PR #875, opened 2026-09-01
Update SAM submodule after container changes

loganharbour:
refs idaholab/moose#33466

PR #878, opened 2026-09-02
SAM submodule update

travismui:
refs [idaholab/moose#33466](https://github.com/idaholab/moose/pull/33466)

Supersedes #875


travismui:
@GiudGiud @loganharbour FYI, I'm flexible with however you want to handle the update. Thanks!

GiudGiud (approved):
thanks for the patch

PR #884, opened 2026-09-11
Update Griffin and BlueCRAB submodules after container changes

gambka:
Refs idaholab/moose#33466


gambka:
@joshuahansel and @GiudGiud. This should fix the devel and apptainer build/compile failures.

GiudGiud:
I turned on the HPC recipes so we can see how these new submodules are doing

gambka:
> I turned on the HPC recipes so we can see how these new submodules are doing

That's fine, but to allow other PRs to get merged without failing recipes we may want to merge this and address the HPC tests in the next submodule update of all apps.

GiudGiud (approved):
test failures to be addressed later

